PR #240, opened 2023-05-29
Support install tools from custom indexes

ischaojie:
We need this feature to install global tools from a custom index. 

``` 
➜ ~ rye install daecli 
Looking in indexes: https://pypi.sapps.douban/simple, https://pypi.org/simple/ 
Collecting daecli
......
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ that were not yet released.
 
 _Unreleased_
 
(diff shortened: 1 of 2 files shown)


ischaojie:
I'm not quite sure if it's best to add changelog to every PR, but it looks like my syntax is in desperate need of improvement. :)

PR #363, opened 2023-07-13
Support shell command in windows

aisk:
The main change is detecting the OSs and shells, and when it's MS's shells (cmd.exe / powershell), using `;` as path seprator. And when it's windows OS, using `venv/Scripts` instead of `venv/bin` as `venv_bin` (https://github.com/pypa/virtualenv/commit/993ba1316a83b760370f5a3872b3f5ef4dd904c1).

Tested on:
- [x] cmd.exe
- [x] powershell
- [x] git bash (mingw) 

Add a new dependency `sysinfo` for get parrent pid and process name. There are some candidates:

- [psutil](https://github.com/rust-psutil/rust-psutil): [do not support windows](https://github.com/rust-psutil/rust-psutil/blob/master/platform-support.md).
- [heim](https://github.com/heim-rs/heim): only support async call.
- sys-info-rs: do not have this api.

The sysinfo is a little too heavy, I think psutil is more suitable if it support windows, but for now it's ok to use sysinfo. Or maybe we can use winapi to call the native library directly?


rye/src/cli/shell.rs
```diff
@@ -36,14 +76,29 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
         .context("failed to sync ahead of shell")?;
 
     let venv_path = pyproject.venv_path();
-    let venv_bin = venv_path.join("bin");
+    let venv_bin = if env::consts::OS == "windows" {
```
aisk, line 79: I think it should be determined in runtime instead of compile time, so using `env::consts::OS == "windows"` instead of `cfg!(windows)`.

Maybe we will using binaries build in windows and runs on linux (wine?).

bluss: Note that this conditional - bin vs Scripts - is also done once already in rye/src/consts.rs

aisk: Updated, thanks for the review!

PR #403, opened 2023-08-04
Add "rye fmt" to reformat with Black

jeanas:
Black is already the de-facto "one true formatter".

### DIFF
--- a/rye/src/bootstrap.rs
+++ b/rye/src/bootstrap.rs
@@ -33,7 +33,7 @@ pub const SELF_PYTHON_TARGET_VERSION: PythonVersionRequest = PythonVersionReques
     suffix: None,
 };
 
-const SELF_VERSION: u64 = 4;
+const SELF_VERSION: u64 = 5;
 
 const SELF_REQUIREMENTS: &str = r#"
 black==23.7.0


T-256:
+ Ruff team also working on formatter too.

cnpryer:
There's no consensus for these utility commands yet. So I'd personally like to see a discussion opened up for that first at the very least first.

T-256:
> but for now Black is quite close to being a universal standard.

I agree, but Rye is project manager, NOT code analyzer (formatter, linter, type checker).
Let's simply use `rye tools` command for those purpose.

bluss:
Here the user can't choose which version of black is used, if they wanted to. A tool or dev-dependency would allow it.
If the users are going to use formatting or linters etc in their editors, then it's important that they can use the same settings and version there as what rye uses? (That seems like a complicated topic and there is no reason to require all that from any signle PR - but we could think about the direction it's heading?)

Maybe my ideas are at odds with rye's philosophy but I'd like to test the idea (in discussion) of having commands like this - rye **format** or rye **test** - where the user can configure what happens for their project when those commands are run.

One thing that I'd like to have part of it is that such commands can be recursive into the whole workspace (chosen by option).

wu-clan:
Apparently, the ruff team did it, and it caught on fast, do not underestimate any of the outputs.

PR #415, opened 2023-08-13
Add `cpython@3.11.4` support

takano32:


takano32:
Python 3.11.5 have been released.

https://www.python.org/downloads/release/python-3115/

PR #519, opened 2023-12-12
Add default script in pyproject.toml for init

stephan-noel:
Addresses https://github.com/mitsuhiko/rye/issues/319.

Currently, having scripts works as long as you do `rye sync`.  However, `rye init` does not provide a default script in `pyproject.toml` despite creating a default `__init__.py`.  This PR streamlines the developer init workflow to approximate Rust more closely so that the following works: 

1) `rye init my_project`
2) `rye sync` (after `cd my_project`)
3) `rye run hello` (no manual editing of pyproject.toml needed)

The following is now generated based on the project name: 

image

Tested locally on a Windows 11 machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ that were not yet released.
 
 _Unreleased_
 
+- By default a script with the name of the project is now also configured.  #519
+
 - Rye now configures hatchling better in `rye init` so that it works with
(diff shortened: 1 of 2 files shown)


geoqiao:
`rye sync` doesn't works when the project name contain ”-” or “_” like `test_script`.

details:https://github.com/mitsuhiko/rye/issues/531

PR #624, opened 2024-02-09
Undo move to pep440_rs 0.4.0.

mostthingsweb:
pep508_rs 0.2.4 (which transitively depended on pep440_rs 0.4.0) was yanked, but poc-monotrail wasn't updated. So now things fail to build in a different way.


cnpryer:
Maybe we can just bump it upstream (https://github.com/konstin/poc-monotrail/pull/65)

PR #660, opened 2024-02-15
Print help when `rye config` is given without any argument

dsp:
We use clap's ArgGroup to make any combination of --get, --set, --set-int, --set-bool and --unset
required. If not we print the help.
- **config: require any of --get**


mitsuhiko (approved):

sondr3:
This PR broke `rye config --show-path`, it now errors out.

```shell
❯ rye config --show-path
error: one or more required arguments were not provided
```

dsp:
Good catch. #70 fixes this.

PR #697, opened 2024-02-19
Fix missing info in README

kwehmeyer:


j178:
The term "blazing fast" is so overused in the Rust world that I've almost developed an aversion to it.

PR #706, opened 2024-02-20
fix rye config --show-path

dsp:
This fixes an issue where `rye config --show-path` will not work. This
regression was originally introduced in #660.

This PR now makes `--show-path` conflicting with any other flag, and groups action and format flags into an ArgGroup. We added integration tests to continuously ensure the behavior is correct.


mitsuhiko:
Now that we have a basic test infrastructure, could you add a test for this?

dsp:
Yeah will do.

Issue #800, opened 2024-02-27
Adding local dependencies to virtual projects doesn't update pyproject.toml with correct path


ma-sadeghi:
### Steps to Reproduce

```shell
# Clone a repository to be used as a local package
git clone https://github.com/AUTODIAL/AutoEIS/
mkdir proj
cd proj
rye init --virtual
rye add autoeis --path ../AutoEIS/
```

### Expected Result

AutoEIS to be successfully added as a local dependency

### Actual Result

```shell
❯ rye add autoeis --path ../AutoEIS/
Added autoeis @ file:///AutoEIS as regular dependency
Reusing already existing virtualenv
Generating production lockfile: /home/amin/Code/battery-degradation/requirements.lock
error: Distribution not found at: file:///AutoEIS
error: could not write production lockfile for project

Caused by:
    failed to generate lockfile
```

`pyproject.toml` after this command:
```toml
...
dependencies = [
    "autoeis @ file:///AutoEIS",
]
...
```
It seems that `rye` didn't update `pyproject.toml` with the correct path of the local dependency.

### Version Info

```shell
❯ rye --version
rye 0.27.0
commit: 0.27.0 (43ee4fce0 2024-02-26)
platform: linux (x86_64)
self-python: cpython@3.12
symlink support: true
uv enabled: true
```

### Stacktrace

I don't think `rye` crashed, but just in case:

```shell
❯ RUST_BACKTRACE=1 rye add autoeis --path ../AutoEIS/
Added autoeis @ file:///AutoEIS as regular dependency
Reusing already existing virtualenv
Generating production lockfile: /home/amin/Code/battery-degradation/requirements.lock
error: Distribution not found at: file:///AutoEIS
error: could not write production lockfile for project

Caused by:
    failed to generate lockfile

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
   1: rye::lock::generate_lockfile
   2: rye::lock::update_single_project_lockfile
   3: rye::sync::sync
   4: rye::sync::autosync
   5: rye::cli::add::execute
   6: rye::cli::execute
   7: rye::main
   8: std::sys_common::backtrace::__rust_begin_short_backtrace
   9: std::rt::lang_start::{{closure}}
  10: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/ops/function.rs:284:13
[...]

PR #817, opened 2024-02-29
Skip `uv compile` step for URL-based dependencies in `add` command

cnpryer:
Drafting this PR to follow up on #800 debug.

This PR updates the add command to avoid passing URL-based dependencies to `uv compile` in `rye add` command.

Testing: Manual + Snapshots


rye/src/cli/add.rs
```diff
@@ -470,21 +470,8 @@ fn resolve_requirements_with_uv(
         }
         let sources = ExpandedSources::from_sources(&pyproject_toml.sources()?)?;
         sources.add_as_pip_args(&mut cmd);
-        let mut child = cmd
-            .stdin(Stdio::piped())
-            .stdout(Stdio::piped())
-            .stderr(Stdio::piped())
-            .spawn()?;
-        let child_stdin = child.stdin.as_mut().unwrap();
-        writeln!(child_stdin, "{}", req)?;
```
cnpryer, line 479: If I'm following, the idea was to spawn the child process to write requirements directly to through stdin. So we're going through `uv` compile first before resolving specifiers? Interesting.

cnpryer: Hmm. Maybe that's not the entire story. I'm definitely missing something, but I'll come back to this.

rye/src/cli/add.rs
```diff
@@ -468,23 +468,29 @@ fn resolve_requirements_with_uv(
...
-
-        let rv = child.wait_with_output()?;
-        if !rv.status.success() {
-            let log = String::from_utf8_lossy(&rv.stderr);
-            bail!("failed to resolve packages:\n{}", log);
-        }
 
-        let mut new_req: Requirement = String::from_utf8_lossy(&rv.stdout).parse()?;
+        let mut new_req = match req.version_or_url {
+            Some(VersionOrUrl::Url(_)) => req.clone(),
```
cnpryer, line 473: I wanted to look into why the requirement would need to make a trip through `uv` to begin with. I'm sure I'm forgetting something obvious, but this at least demonstrates where the issue might be coming from.

cnpryer:
I believe this was expected, but I tried recreating this issue as a `uv` tests since there are stdin tests over there for `pip compile`. So far I can't reproduce it.

```rs
/// Resolve with PROJECT_ROOT path from a `requirements.in` file on stdin when
/// passed a path of `-`.
#[test]
fn compile_requirements_in_stdin_path() -> Result<()> {
    let context = TestContext::new("3.12");
    let requirements_in = context.temp_dir.child("requirements.in");
    requirements_in.write_str("maturin_editable @ file:///${PROJECT_ROOT}/../../scripts/editable-installs/maturin_editable")?;

    uv_snapshot!(Command::new(get_bin())
    .arg("pip")
    .arg("compile")
    .stdin(fs::File::open(requirements_in)?)
    .arg("-"), @r###"
    success: true
    exit_code: 0
    ----- stdout -----
    # This file was autogenerated by uv via the following command:
    #    uv pip compile -
    maturin-editable @ file:///${PROJECT_ROOT}/../../scripts/editable-installs/maturin_editable

    ----- stderr -----
    Resolved 1 package in [TIME]
    "###);

    Ok(())
}
```

cnpryer:
I'd like to read into URL-based dependencies more before marking this as ready.

In the meantime I'll share my thoughts so far. Since URL-based dependencies are direct references, I'd assume that `uv` would always emit the same requirements string each time. So if that's the case, I'd think doing this here, where we are just skipping that trip for URL-based dependencies, could work.

I'll give it some more thought.

rye/src/cli/add.rs
cnpryer: See [comment](https://github.com/astral-sh/rye/pull/817#issuecomment-1975280848)

bluss: As the code is here, you skip using the result from uv for URLs and that makes sense (If I understand the code correctly) - but should we then also skip using uv at all? I don't know if it's important for the add command do some kind of validation of the arguments like this (checking that they exist)

cnpryer: Yup we seem to be on the same page. Initially it seems like we can just skip it since these are direct references, but I want to understand the big picture a little better.

Thanks for looking at this!

rye/tests/test_add.rs
```diff
@@ -163,3 +163,52 @@ fn test_add_flask_wrong_venv_exported() {
...
+    space.init("my-project");
+    let fake_venv = space.project_path().join("fake-venv");
+    fs::create_dir_all(&fake_venv).unwrap();
+
+    let wheel_path = get_fixtures_dir()
+        .join("wheels")
+        .join("rye_devtools-1.0.0-py3-none-any.whl");
+    rye_cmd_snapshot!(space.rye_cmd().arg("add").arg("rye_devtools").arg("--path").arg(wheel_path).env("VIRTUAL_ENV", fake_venv.as_os_str()), @r###"
+    
+    "###);
```
cnpryer, line 179: Need to:
1. Add fixtures filter
2. Maybe use a different wheel for now (see below)

```
program: rye
args:
  - add
  - rye_devtools
  - "--path"
  - /Users/chrispryer/github/rye/rye/tests/fixtures/wheels/rye_devtools-1.0.0-py3-none-any.whl
env:
  RYE_HOME: /Users/chrispryer/github/rye/target/debug/rye-test-dir/home
  __RYE_UV_EXCLUDE_NEWER: "2023-11-18T12:00:00Z"
  VIRTUAL_ENV: /var/folders/x5/94x0_4tj6hz4682vh8l4fv1m0000gn/T/.rye-tests---01Jgvf/project/fake-venv
  UV_CACHE_DIR: /var/folders/x5/94x0_4tj6hz4682vh8l4fv1m0000gn/T/.rye-tests---01Jgvf/uv-cache
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
+new results
────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────
          0 │+success: false␊
          1 │+exit_code: 1␊
          2 │+----- stdout -----␊
          3 │+Initializing new virtualenv in [TEMP_PATH]/project/.venv␊
          4 │+Python version: cpython@3.12.2␊
          5 │+Added rye_devtools @ file:///Users/chrispryer/github/rye/rye/tests/fixtures/wheels/rye_devtools-1.0.0-py3-none-any.whl as regular dependency␊
          6 │+Reusing already existing virtualenv␊
          7 │+Generating production lockfile: [TEMP_PATH]/project/requirements.lock␊
          8 │+␊
          9 │+----- stderr -----␊
         10 │+  × No solution found when resolving dependencies:␊
         11 │+  ╰─▶ Because only httpx[socks]<0.26.0 is available and rye-devtools==1.0.0␊
         12 │+      depends on httpx[socks]>=0.26.0, we can conclude that␊
         13 │+      rye-devtools==1.0.0 cannot be used.␊
         14 │+      And because only rye-devtools==1.0.0 is available and you require␊
         15 │+      rye-devtools, we can conclude that the requirements are unsatisfiable.␊
         16 │+␊
         17 │+      hint: Pre-releases are available for httpx[socks] in the requested␊
         18 │+      range (e.g., 1.0.0b0), but pre-releases weren't enabled (try:␊
         19 │+      `--prerelease=allow`)␊
         20 │+error: could not write production lockfile for project␊
         21 │+␊
         22 │+Caused by:␊
         23 │+    failed to generate lockfile␊
────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

cnpryer:
I'll mark this as ready, but I want to go and review some issues that have been submitted. I think there might be a few this can either begin to address or close completely.

rye/src/cli/add.rs
```diff
@@ -468,23 +468,29 @@ fn resolve_requirements_with_uv(
...
-
-        let rv = child.wait_with_output()?;
-        if !rv.status.success() {
-            let log = String::from_utf8_lossy(&rv.stderr);
-            bail!("failed to resolve packages:\n{}", log);
-        }
 
-        let mut new_req: Requirement = String::from_utf8_lossy(&rv.stdout).parse()?;
+        // we can skip the `uv compile` step for dependencies with direct references (URLs/paths)
+        let mut new_req = if matches!(req.version_or_url, Some(VersionOrUrl::Url(_))) {
```
charliermarsh, line 473: Could this be simplified to just:

```
if matches!(req.version_or_url, Some(VersionOrUrl::Url(_))) {
    continue;
}
```

At the very top of the loop (line 452)?

cnpryer:
A little premature to mark as ready. I need to [tweak this](https://github.com/astral-sh/rye/pull/817#discussion_r1513801721), add a `--git` test, and tag some related issues.

This might warrant a CHANGELOG entry as well.

PR #897, opened 2024-03-16
Fix `rye add m --path ./m` panic on windows

j178:
`Url::from_file_path` returns Err on Windows if the prefix is not a disk prefix (e.g. C:), we can work around this by using `Url.set_path`.

## Steps to reproduce
```
rye init --build-system setuptools hello
cd hello
rye add m --path ./m
```


bluss:
In some configurations, rye supports adding relative paths as dependencies, and I think with this change that's then impossible yo do from the cli on windows  (and apparently it was already impossible due to this bug)

j178:
@bluss Indeed, I switched to using `Url.set_path` as a workaround for the `Url::from_file_path` panic, and this also made it working on Windows.

bluss:
thanks for that @j178

PR #921, opened 2024-03-22
Refuse to remove the target directory if it is not a python installation

j178:
This protects `rye fetch 3.12 -f --target-path /` from removing the whole root directory.


bluss:
maybe this interaction between --target_path and force can be mentioned on the help for the target_path option? That it will be deleted if you use force.

bluss:
Someone on twitter said rye fetch cpython@3.12 --target-path ~/.local

That makes me think the force and target path interaction needs even stronger protections against mishaps. This PR is a good first step of course.

PR #928, opened 2024-03-23
Update pep508_rs and pep440_rs

### LINKED ISSUE #912: Can't add path dependencies in virtual project
### Steps to Reproduce

```
mkdir relativebug
cd relativebug
rye init a
rye init main
cd main
rye add a --path ../a
# Added a @ file:///home/user/ryes/relativebug/main/../a as regular dependency

cd ..
rye init main2 --virtual
cd main2
rye add a --path ../a
# error: Failed to run uv compile error: Distribution not found at: file:///a
```

### Expected Result

In rye 0.29, it adds the dependency `"a @ file:///${PROJECT_ROOT}/../a",`  to the virtual project (here "main2").
Note that a virtual project like main2 or a build_system=pdm project behave the same way, they both use the PROJECT_ROOT relative paths.

### Actual Result

Error message

```
# error: Failed to run uv compile error: Distribution not found at: file:///a
```

### Version Info

rye 0.30.0
commit: 0.30.0 (c7cbdd955 2024-03-19)
platform: linux (x86_64)
self-python: cpython@3.12.2
symlink support: true
uv enabled: true

### Stacktrace

_No response_

bluss:
Update pep508_rs and pep440_rs crates

Main reason is to use VerbatimUrl from pep508_rs and try to fix issues around file:///${PROJECT_ROOT}/etc dependencies.

Note that monotrail-utils is still using the older pep508_rs.

Fixes #912

### DIFF
--- a/rye/src/cli/init.rs
+++ b/rye/src/cli/init.rs
@@ -701,7 +701,8 @@ fn import_requirements_file(
     data.requirements.iter().for_each(|x| {
         requirements
             .entry(x.requirement.name.to_string())
-            .or_insert(format_requirement(&x.requirement).to_string());
+            // todo: will to_string this work?
+            .or_insert(x.requirement.to_string());
     });
     Ok(())
 }
--- a/rye/src/cli/init.rs
+++ b/rye/src/cli/init.rs
@@ -25,8 +25,8 @@ use crate::platform::{
 use crate::pyproject::BuildSystem;
 use crate::sources::py::PythonVersionRequest;
 use crate::utils::{
-    copy_dir, escape_string, format_requirement, get_venv_python_bin, is_inside_git_work_tree,
-    CommandOutput, CopyDirOptions, IoPathContext,
+    copy_dir, escape_string, format_requirement, format_requirement_v030, get_venv_python_bin,
+    is_inside_git_work_tree, CommandOutput, CopyDirOptions, IoPathContext,
 };
 
 /// Initialize a new or existing Python project with Rye.
@@ -701,8 +701,7 @@ fn import_requirements_file(
     data.requirements.iter().for_each(|x| {
         requirements
             .entry(x.requirement.name.to_string())
-            // todo: will to_string this work?
-            .or_insert(x.requirement.to_string());
+            .or_insert(format_requirement_v030(&x.requirement).to_string());
     });
     Ok(())
 }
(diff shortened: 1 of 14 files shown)


bluss:
This is in an attempt to solve #912. In this case by using the real absolute path as the Url and the ${PROJECT_ROOT}/../relative path as the "given" part of the URL, let's see how that works out.

Remains to be done - test the monotrail dependent import.
Add additional tests for #912 - or do that separately?

rye/src/cli/init.rs
```diff
@@ -700,7 +700,8 @@ fn import_requirements_file(
     data.requirements.iter().for_each(|x| {
         requirements
             .entry(x.requirement.name.to_string())
-            .or_insert(format_requirement(&x.requirement).to_string());
+            // todo: will to_string this work?
+            .or_insert(x.requirement.to_string());
```
bluss, line 704: tested - the missing format_requirement here messes up `${VARIABLES}` in requirements files. It's not called in the PR because of the pep440 version divergence. One could depend on both new and old, any solution feels gross here.

bluss:
Now it's working, but it certainly does not feel solid around VerbatimUrl and ${}'s.

bluss:
rebased

bluss:
Feedback on this is welcome. Updating crates is ok? Fixing relative paths is ok or a bad idea? We could go other ways, like adopting another convention for relative paths than the `${PROJECT_ROOT}` stuff here?

charliermarsh:
Can you explain why the original failure is happening? I.e., why is `PROJECT_ROOT` problematic in a virtual project?

bluss:
The bug is just general to rye add with relative paths.

The default `rye init` project is *not* virtual and build system is hatchling. In this configuration, currently, Rye does not support relative paths with git add. It just makes them absolute. So, no bug with the default project.

This bug shows up in for example `rye init --virtual` and `rye init --build-system pdm` projects where rye add thinks it supports relative path dependencies, and fails to construct the URL for it.

So we could say that ${PROJECT_ROOT} is always problematic for rye add currently - it seems to work in rye sync though.

This PR is my attempt to fix this in a "minimal" way. It's not great that the PR mixes two things - updating deps and fixing bugs in add, but when I wrote this I didn't want to leave stuff in a half broken state.

bluss:
There's a lot more we'd like to do with relative paths, but I need some buy in from @mitsuhiko I feel. It would be great to adopt some ideas from uv where you have done the hard work of finding out what's interoperable with pip and what is not - see PR https://github.com/astral-sh/uv/pull/3682

bluss:
rebased and dropped the changelog update because the newest changelog seemed to be autogenerated

bluss:
Will close

PR #942, opened 2024-03-25
Fix rye list overwrites .venv

ischaojie:
fixes https://github.com/astral-sh/rye/issues/940, If the project was not created by rye, you can still use `rye list`.

```
➜  test mkdir test1
➜  test cd test1
➜  test1 pdm init -n
Creating a pyproject.toml for PDM...
Virtualenv is created successfully at /Users/zhezhezhu/projects/test/test1/.venv
Project is initialized successfully

PDM 2.10.0 is installed, while 2.12.4 is available.
Please run `pdm self update` to upgrade.
Run `pdm config check_update false` to disable the check.
➜  test1 pdm add jinja2
Adding packages to default dependencies: jinja2
🔒 Lock successful
Changes are written to pyproject.toml.
Synchronizing working set with resolved packages: 2 to add, 0 to update, 0 to remove

  ✔ Install markupsafe 2.1.5 successful
  ✔ Install jinja2 3.1.3 successful

🎉 All complete!


PDM 2.10.0 is installed, while 2.12.4 is available.
Please run `pdm self update` to upgrade.
Run `pdm config check_update false` to disable the check.
➜  test1 rye list
Detected outdated rye internals. Refreshing
Bootstrapping rye internals
Found a compatible Python version: cpython@3.12.0
jinja2==3.1.3
markupsafe==2.1.5

```


dsp:
I am a bit late to the party, but I actually don't like this change. It subtlety breaks the constraint that you always build a venv instance from Uv to ensure the tool exists locally. I'd rather have an 'assume_venv' or something on Uv. The way Uv and UvVenv is designed, is such that we always ensure a Uv binary is present. This PR breaks this.

bluss:
@dsp Should it be a new bug? IMO the feature of "opening" any other venv is not important, the only important thing is to not overwrite venvs with rye list. "rye list" is an interrogator command that should never be destructive (and preferably never to any modifications - just IMO.)

PR #943, opened 2024-03-25
Bump minor deps

kevherro:
This change bumps a handful of dependencies to their next minor versions.

`pep440_rs` and `pep508_rs` are also candidates, but both require non-trivial code changes. I left those out for now.


bluss:
pep508 is updated in #928

PR #959, opened 2024-03-28
Fix verbose/quiet for rye lint and format

bluss:
These commands produced `ruff --verbose check`  (same for --quiet and same for format) but the verbose/quiet need to come after the subcommand. The command would produce an error (file `check`/`format` does not exist), now it succeeds.

Fixes #956

PR #977, opened 2024-04-01
rye fmt ruff extra args error (degrade)

sio-funmatsu:
Fixed the following error:

```sh
$ rye fmt -q -- --select I --fix
error: unexpected argument '--select' found

  tip: to pass '--select' as a value, use '-- --select'

Usage: ruff format <--verbose|--quiet|--silent> [FILES]...

For more information, try '--help'.
```


bluss:
Ok, I'm sorry about that, but this change as written would reintroduce the bug fixed here, right? #959

PR #978, opened 2024-04-01
Use uv in rye build when enabled

bluss:
The `build` package supports uv since version 1.2.0 - it can use either uv or pip as installer.

The main way to use it would be to install `build[uv]` but rye already has uv installed separately, build will find it on the path as a fallback, so this can work for rye.

build is implicitly tested by test_publish.

Relates to #668

### DIFF
--- a/rye/src/cli/build.rs
+++ b/rye/src/cli/build.rs
@@ -104,6 +104,10 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
             build_cmd.arg("--sdist");
         }
 
+        if output == CommandOutput::Verbose {
+            build_cmd.arg("--verbose");
+        }
+
         if output == CommandOutput::Quiet {
             build_cmd.stdout(Stdio::null());
             build_cmd.stderr(Stdio::null());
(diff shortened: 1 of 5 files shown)


rye/src/cli/build.rs
```diff
@@ -75,14 +78,23 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
...
             .arg(&out)
             .arg(&*project.root_path());
 
+        if use_uv {
+            // we need to ensure uv is available to use without installing it into self_venv
+            let uv = UvBuilder::new().with_output(output).ensure_exists()?;
+            build_cmd.arg("--installer=uv");
+            build_cmd.env("PATH", uv.uv_bin().parent().ok_or_else(||
+                anyhow!("Could not find uv binary in self venv: empty path")
+            )?);
```
bluss, line 95: This PATH upgrade needs review or comments before this is done. In this shape, it works for me. And it's faster too. :slightly_smiling_face:

bluss: The question is, is it needed to expose the rest of the PATH variable to the build?

bluss: Now I've gone with prepend to $PATH as the solution instead, which should be the normal kind of solution.

PR #986, opened 2024-04-02
DOC: simplify wording for rye sync after rye add

willingc:
This PR improve text in the doc so a new user of rye doesn't miss the need to `rye sync` after adding a dependency to actually install it. I skimmed past the original text so hopefully this will help the next reader.


santigandolfo:
Isn't `sync` run automatically after `add` or is that just for `uv`? Should that be mentioned?

charliermarsh (approved):

charliermarsh:
> Isn't sync run automatically after add or is that just for uv? Should that be mentioned?

Yeah, although uv isn't the default yet so I think it's better not to introduce that complexity at this point in the docs.

PR #987, opened 2024-04-02
DOC: correct black run command and refine text

willingc:
This PR corrects the run command which will error if black is not given a directory to check. Refined wording to reinforce "add-install-run" steps.

### DIFF
--- a/docs/guide/basics.md
+++ b/docs/guide/basics.md
@@ -98,12 +98,13 @@ rye remove flask
 ## Working with the Project
 
 To run executables in the context of the virtualenv you can use the `run` command.  For
-instance if you want to use `black` you can add and run it like this:
+instance, if you want to use `black` you can add, install, and run it on the current
+directory like this:
 
 ```
 rye add black
 rye sync
-rye run black
+rye run black .
 ```
 
 If you want to have the commands available directly you will need to activate the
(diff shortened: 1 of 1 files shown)


docs/guide/basics.md
```diff
@@ -98,12 +98,13 @@ rye remove flask
 ## Working with the Project
 
 To run executables in the context of the virtualenv you can use the `run` command.  For
-instance if you want to use `black` you can add and run it like this:
+instance, if you want to use `black` you can add, install, and run it on the current
```
bluss, line 101: should we say sync instead of install? Because "rye install" is something else (that is also black related, potentially), just a thought, not sure.

willingc: @bluss I suppose it could be either "install or sync". The section above talks about `rye sync` installing the dependencies that have been added. If read by a new user, the behavior is what I want to understand. Add the black dependency (`rye add black`, install the black package (`rye sync`), and run black on the current directory (`rye run black .`). 

What does `rye install` do?

willingc: Ah...I see now `rye install` or `rye tools install` will install into the global space instead of the virtual environment. 

Maybe `you can add, sync the virtual environment which will install the dependency into it, and run...`

PR #1001, opened 2024-04-08
doc: Typos and wording

emarsden-iso:
Noticed a couple of typos, and places where I thought the wording could be made a bit clearer.


charliermarsh (approved):
These look reasonable to me, thanks.

PR #1015, opened 2024-04-13
Support uv --override option (#668)

idlsoft:
This change will read an `override-dependencies` array from `[tool.rye]` and propagate it as an `--override` option to `uv`

Example:
```
[project]
...
dependencies = ["numpy==1.26.4"]

[tool.rye]
managed = true
override-dependencies = [ "numpy==1.26.1" ]
```


bluss:
I'm not a fan of expanding scope, so don't do that, but rye add can add deps, dev-deps and excluded deps. Extending the pattern it seems like it would fit that rye add can add overrides as well, which could be a later PR.

rye/src/lock.rs
```diff
@@ -397,6 +419,7 @@ fn generate_lockfile(
             }
         };
 
+        // here ??
```
bluss, line 422: now obsolete comment?

rye/src/lock.rs
```diff
@@ -194,6 +197,21 @@ pub fn update_workspace_lockfile(
     Ok(())
 }
 
+fn maybe_write_requirements_to_temp(requirements: &HashSet<Requirement>) -> Result<Option<NamedTempFile>, Error> {
+    let over_temp_file: Option<NamedTempFile> = match requirements.is_empty() {
```
bluss, line 201: If this prefers `if` instead of `match`  (I don't mind match bool, but if is less surprising) and removes the over_temp_file binding, the whole function body is just the if-else, then this function becomes simple.

idlsoft: good point

rye/src/lock.rs
idlsoft: yes :)

idlsoft:
> I'm not a fan of expanding scope, so don't do that, but `rye add` can add deps, dev-deps, optional deps and excluded deps. Extending the pattern it seems like it would fit that rye add can add overrides as well, which could be a later PR.

Done

rye/src/lock.rs
```diff
@@ -425,13 +450,17 @@ fn generate_lockfile(
...
+                overrides_file_in,
                 &requirements_file,
                 lock_options.pre,
                 env::var("__RYE_UV_EXCLUDE_NEWER").ok(),
                 upgrade,
                 keyring_provider,
             )?;
     } else {
+        if overrides_file_in.is_some() {
+            bail!("dependency overrides require the uv backend");
```
bluss, line 462: can this maybe be a warning? Don't know

bluss (approved):
lgtm. I can review, but note I'm just a contributor and can't do more in this repo.

rye/src/lock.rs
idlsoft: it would ignore an explicit instruction to overwrite a dependency, I feel like it should be treated as an error.

idlsoft:
> lgtm. I can review, but note I'm just a contributor and can't do more in this repo.

I don't think I can ask for reviewers. Your name was there because you've commented earlier.
Thx :)

rye/src/lock.rs
bluss: It's a question of design, I don't know, on one hand the project is a pyproject.toml file which can be interpreted by standard python tools, not just Rye. And those tools would ignore everything inside `tool.rye` anyway. Maybe error when rye is managing it is good yeah.

PR #1016, opened 2024-04-13
Add keyring support for uv

emarsden-iso:
`uv` has support for keyrings, as long as the option is `--keyring-provider=subprocess`. But this isn't exposed in the `rye` CLI.

Some things I haven't addressed, happy to discuss:

1. I haven't implemented it for piptools. I expect it's possible, in fact more features could be implemented using pip-compile's `--pip-args` argument, but I think the plan is to move to `uv` as the default going forward, so I don't know how much desire there is to extend the piptools backend. But if it's needed, I can do it, either in this PR or as followup work.

2. In practice, I'd expect that people would want to always pass this `--keyring-provider` value for all calls, and it does pollute the code a bit, passing round the CLI options. I think another way to go could be to put it in the global config and/or the project TOML, and just read it from there. (I'd actually prefer this interface for my own uses.) But I wanted to stay close to the existing CLI of `uv` and `pip`, so stuck with that interface. Again, happy to iterate.

# Test

From a `rye init` project:

```
❯ ../rye/target/debug/rye lock
Generating production lockfile: /usr/local/google/home/emarsden/src/rye-proj/test-proj/requirements.lock
error: Failed to build editables
  Caused by: Failed to build editable: file:///usr/local/google/home/emarsden/src/rye-proj/test-proj
  Caused by: Failed to install requirements from build-system.requires (resolve)
  Caused by: No solution found when resolving: hatchling
  Caused by: HTTP status client error (401 Unauthorized) for url (https://europe-python.pkg.dev/isolabs-builds/isolabs-virtual/simple/hatchling/)
error: could not write production lockfile for project

Caused by:
    Failed to run uv compile /tmp/.tmplGQmy9/requirements.txt. uv exited with status: exit status: 2

❯ ../rye/target/debug/rye lock --keyring-provider=subprocess
Generating production lockfile: /usr/local/google/home/emarsden/src/rye-proj/test-proj/requirements.lock
Generating dev lockfile: /usr/local/google/home/emarsden/src/rye-proj/test-proj/requirements-dev.lock
Done!
```


rye/src/lock.rs
```diff
@@ -136,6 +143,7 @@ pub fn update_workspace_lockfile(
     output: CommandOutput,
     sources: &ExpandedSources,
     lock_options: &LockOptions,
+    keyring_provider: KeyringProvider,
```
charliermarsh, line 146: Should this be on `LockOptions`, rather than passed in as a standalone argument?

charliermarsh (approved):

rye/src/lock.rs
charliermarsh: (I decided: no, because it's also needed for `sync` -- which does take `LockOptions`, but for unrelated reasons.)

emarsden-iso:
@charliermarsh thanks for taking a look and fixing things up, looks good to merge now?

charliermarsh:
Thanks for the nudge, I honestly though it had merged.

PR #1034, opened 2024-04-22
Include sources in lockfile

GeorchW:
Including the sources in the lockfile has the advantage of being able to install the dependencies with `pip` or `uv` directly from the lockfile.

From the code it seems like there was a conscious decision to do it differently, but I'm not sure what the reasoning was back then and whether it's still applicable.
I'm absolutely open for discussions!


GeorchW:
Nvm, I just wasn't aware of the `--with-sources` option.

PR #1053, opened 2024-04-26
Flip uv to the default Rye backend

charliermarsh:
## Summary

I think `uv` is now stable enough to serve as the default backend for Rye. Users can explicitly opt-out with, e.g., `rye config --set-bool behavior.use-uv=false`, to fall back to the `pip-tools` backend.

Suggested here: https://twitter.com/mitsuhiko/status/1783853343502160290

### DIFF
--- a/docs/guide/config.md
+++ b/docs/guide/config.md
@@ -81,9 +81,10 @@ force-rye-managed = false
 # virtual environments.
 global-python = false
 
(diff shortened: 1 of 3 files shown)


mitsuhiko (approved):

samuelcolvin (approved):
👍

PR #1054, opened 2024-04-27
Fix Rye not using user-chosen toolchain as default during installation

pjdon:
Fixes https://github.com/astral-sh/rye/issues/1024

Currently:
* When installation calls  `ensure_self_venv_with_toolchain` to bootstrap rye internals it would pass the `toolchain_version_request` variable as the version to install
* If the `RYE_TOOLCHAIN_VERSION` env var is not set, then the value of `toolchain_version_request` remains None
* During installation, `prompt_for_default_toolchain` is called to prompt the user for a version to select, asking the user `"Which version of Python should be used as default toolchain?"`
* This function stores the user's selection in the `config_doc` TOML configuration and returns nothing
* The user's toolchain version selection is thus never passed to the internals bootstrapping call

Changes:
* This fix makes `toolchain_version_request` mutable so it can be updated by the user's input
* Function `prompt_for_default_toolchain` returns the resolved version that the user wants, which may be the input they typed or the default value
* The result is stored in `toolchain_version_request` so it can be used during the internals bootstrapping

### DIFF
--- a/rye/src/cli/rye.rs
+++ b/rye/src/cli/rye.rs
@@ -624,12 +624,17 @@ fn perform_install(
         // can fill in the default.
         if !matches!(mode, InstallMode::NoPrompts) {
             if toolchain_path.is_none() {
-                toolchain_version_request = Some(prompt_for_default_toolchain(
+                // Prompt the user for the default toolchain version.
+                let user_version_request = prompt_for_default_toolchain(
                     toolchain_version_request
                         .clone()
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

charliermarsh:
Thanks, this makes sense. I just tweaked the code a bit to respect the `RYE_TOOLCHAIN_VERSION` if it _is_ set. I could see an argument either way... but I think the set env var overriding a prompted "default" version is reasonable.

PR #1070, opened 2024-05-10
Allow to generate lockfiles with hashes when using uv.

mvaled:
This PR add the option `--generate-hashes` to both `rye sync` and `rye lock`.  It will generate lock files with hashes if using `uv`.


charliermarsh (approved):

mvaled:
Hi @charliermarsh 

Thanks for the quick review.  I fixed the failing tests.

mvaled:
Hi @charliermarsh 

The Test (Window) is failing due to an access denied.

```
         11 │+Caused by:
         12 │+    Access is denied. (os error 5) at path "C:/Users/RUNNER~1/AppData/Local/Temp/.tmplSh3zx"
```

I think this is probably not related to the PR.

PR #1074, opened 2024-05-12
Workspace members: Added symbolic link support

LLouice:


zanieb:
Hi! Can you say a bit more about why this change is important or link to some prior discussion?

PR #1075, opened 2024-05-12
Bump boostrap ruff to 0.4.4

davfsa:
Noticed ruff was a bit old and was missing a lot of rules, leading to a lot of "unused noqa".

I unfortunately dont have the rust knowledge to improve this, but it would be nice if we could update ruff independently of rye and still run it using `rye lint/fmt`.

Thanks for this amazing tool! Gave it a shot today and really really enjoyed playing around with it and testing it out. Currently porting some of my projects to use it!

### DIFF
--- a/rye/src/bootstrap.rs
+++ b/rye/src/bootstrap.rs
@@ -33,7 +33,7 @@ pub const SELF_PYTHON_TARGET_VERSION: PythonVersionRequest = PythonVersionReques
     suffix: None,
 };
 
-const SELF_VERSION: u64 = 18;
+const SELF_VERSION: u64 = 19;
 
 const SELF_REQUIREMENTS: &str = r#"
 build==1.2.1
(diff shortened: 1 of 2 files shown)


davfsa:
The last tests failure is insanely weird and I'm not sure if its expected . Ill leave it without change and let someone else have a look at it, because I cant reproduce it locally and I'm not sure what other file it could be talking about

https://github.com/astral-sh/rye/actions/runs/9054124712/job/24873624010#step:7:206

bluss:
Yes, space.init looks like it uses regular "rye init" which creates src/packagename/__init__.py already. I think the new output is expected and can just be confirmed then?

davfsa:
Thanks for the extra pair of eyes. I had a suspicion it had something to do with that, but was not 100% sure. Made the test account for that too

charliermarsh (approved):

charliermarsh:
Thanks!

Issue #1086, opened 2024-05-15
How to test a virtual package?


dennisrall:
How do I need to setup my project to be able to run tests for a virtual package? At the moment I am trying the standard way:
```
├── src
    ├── virt_package
        ├── __init__.py
        ├── main.py
├── tests
    ├── virt_package
        ├── test_main.py
└── pyproject.toml
```

How would I import a method `foo` from the `main.py` in the `test_main.py`? Is this even possible in virtual packages as the package itself is not installed?

dennisrall:
It was just a confusion about what virtual packages are

jennydaman:
I guess I am confused too and the docs do need more elaboration on this topic.

https://rye.astral.sh/guide/docker/ <-- the guide here seems to suggest that projects which you intend to use Docker with are recommended to be virtual. It explains what the effects are but not why it is desirable.

dennisrall:
As far as I know if your project is so big that you need tests it shouldn't be a virtual project. And therefore you have to adapt the Dockerfile. But I agree that this is still confusing

jennydaman:
In case anyone else stumbles upon this thread, here's what I am doing:

For development on-the-metal, the rye project is not a virtual project, so that I can define my pytest tests in `tests/`.

My `Dockerfile` is what the docs recommend to use for virtual projects, but with an extra `sed '/.../d'` command to delete the line that installs the project as a package.

```Dockerfile
FROM python:3.12.3-alpine

WORKDIR /app
COPY requirements.lock ./
RUN sed -i' ' -e '/-e file:\./d' requirements.lock \
    && env PYTHONDONTWRITEBYTECODE=1 pip install --no-cache-dir -r requirements.lock

COPY src .
CMD ["fastapi", "run", "main.py"]
```

PR #1087, opened 2024-05-16
docs: add hint to add the src folder to the python path for virtual p…

dennisrall:
I had some problems testing virtual projects, see  #1086.
I added a hint on how to add the `src` folder to the python path with pytest.

### DIFF
--- a/docs/guide/virtual.md
+++ b/docs/guide/virtual.md
@@ -26,6 +26,17 @@ When syncing the project itself is never installed into the virtualenv as it's n
 considered to be a valid package.  Likewise you cannot publish virtual packages to
 PyPI or another index.
 
+### Run tests outside of application code
+
+If you tests are separated from your application code (typically by putting them in a
+`tests` folder next to the `src` folder) you need to add the `src` folder to the python
+path. With pytest you can achieve this with the following section in your `pyproject.toml`
+file:
+```toml
+[tool.pytest.ini_options]
+pythonpath = "src"
+```
+
 ## Limitations
 
 Virtual projects can not have optional dependencies.  These even if declared are not


bluss:
When you say virtual, you don't mean the same thing as "rye init --virtual" I guess (from context), or am I wrong?

bluss:
Maybe I am wrong and you mean that. But why does one have a "src" directory for a rye init --virtual project? That seems like a contradiction.

dennisrall:
Yes I mean projects you get by running "rye init --virtual". But maybe then I am wrong about them. The docs about [virtual Projects](http://rye-up.com/guide/virtual/) introduce them for just running python packages. Seems to me a bit like `rye install`? But the [docker documentation](http://rye-up.com/guide/docker/#container-from-source) copies the `src` folder from the virtual project to the Dockerfile.

So, do virtual projects contain code or not?

bluss:
I would think of it like this - virtual projects are there if you want rye to manage a python environment (.venv) with a special list of packages (dependencies) without developing anything that looks like a python package (like those that are published to pip).

Code takes many shapes - probably there is code in a virtual project, but not necessarily. But if you want to have a python module that is imported from other scripts, that sounds like a python module that should be installed into the venv, like how it's done in a regular rye project (not virtual)

So if you feel a need to change PYTHONPATH, that sounds like you just want those python modules in the environment and a regular rye init --lib project seems like the right thing to use?  With that said, I don't have any workflow that uses docker.

dennisrall:
I like the perspective of viewing the `tests` directory as a separated package where you want to install your package. I think then a normal package is better suited for this.

PR #1091, opened 2024-05-18
Discover cosmo-ified (`.com`) binaries on Windows

mataha:
Executables compiled with [Cosmopolitan](https://justine.lol/cosmopolitan/index.html) libc runtime have a `.com` extension on Windows. This includes cosmo-ified Python binaries.

Example [here](https://github.com/astral-sh/rye/files/15366179/python.zip).

### DIFF
--- a/rye/src/utils/mod.rs
+++ b/rye/src/utils/mod.rs
@@ -166,7 +166,7 @@ pub fn is_executable(path: &Path) -> bool {
     }
     #[cfg(windows)]
     {
-        ["exe", "bat", "cmd"]
+        ["com", "exe", "bat", "cmd"]
             .iter()
             .any(|x| path.with_extension(x).is_file())
     }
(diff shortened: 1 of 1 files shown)


zanieb:
Hi! Did you confirm that these work once discovered?

zanieb:
And is there a reason they can't use the system standard executable suffix?

mataha:
> Hi! Did you confirm that these work once discovered?

Python works, which is good enough for me. I was hoping that scripts would work as well, but these report "program not found" as `std::process::Command` is used under the hood with [all of its shortcomings](https://github.com/BurntSushi/ripgrep/blob/master/crates/cli/src/decompress.rs#L427-L445).

> And is there a reason they can't use the system standard executable suffix?

It's to distinguish them from the standard PE executable format. That said, Windows itself bundles some `.com` binaries for compatibility purposes (`more` being the most egregious example) even though they're PE.

mataha:
Thanks!

PR #1143, opened 2024-06-07
Fixed bug 1136

Zander-1024:
Redirect to stdout to facilitate filtering using grep.

https://github.com/astral-sh/rye/issues/1136


zanieb:
Can you share a few examples demonstrating exactly what goes to stdout and stderr with this change? e.g. do error messages still go to stderr?

Zander-1024:
image
I think that's pretty clear.

mitsuhiko:
This is definitely the wrong fix.  The right fix is to drop the guard in the list command.  I will make a replacement PR.

PR #1148, opened 2024-06-09
Imply not using cache when invoking sync --update-all

flyaroundme:
Hi! This is an attempt to fix https://github.com/astral-sh/rye/issues/1128 by basically passing the `--no-cache`  to `uv` command on lockfile generation.
This should work exactly the same (https://github.com/astral-sh/uv/blob/8ae5c2aee3fa22ac7d11337688e4c1bee3748fe3/crates/uv-cache/src/cli.rs#L17) as adding `UV_NO_CACHE=true` as environment variable as the issue author tried in order to get the desired behaviour.

### DIFF
--- a/rye/src/uv.rs
+++ b/rye/src/uv.rs
@@ -66,7 +66,7 @@ impl UvCompileOptions {
 
         match self.upgrade {
             UvPackageUpgrade::All => {
-                cmd.arg("--upgrade");
+                cmd.arg("--upgrade").arg("--no-cache");
             }
             UvPackageUpgrade::Packages(ref pkgs) => {
                 for pkg in pkgs {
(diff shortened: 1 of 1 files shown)


zanieb:
Interesting. Thanks for putting up a fix. You should be able to use `--refresh` instead of `--no-cache`, the latter will disable writes to the cache as well which is overkill. Not entirely sure if there's a better approach than that though without looking into the original issue further.

zanieb:
I'm still a bit confused about why this is necessary though.

flyaroundme:
> I'm still a bit confused about why this is necessary though.

Got you, no problem. The use-case is described in the issue, so I thought that made sense

zanieb:
But if there's a new release, uv should upgrade to it without refreshing — unless it's been released in the last 10 minutes (which is the TTL for cache of the available releases). It might make more sense to just expose the `--refresh` flag in Rye. I think we'll need a clearer picture of the problem before it makes sense to change the uv invocation.

PR #1154, opened 2024-06-11
shell completion argument is `power-shell` not `powershell`

jimkring:
should be:
`rye self completion -s power-shell`

not:
`rye self completion -s powershell`

This can be verified via the help:

`rye self completion --help`


charliermarsh:
Thanks! However we fixed the serialized representation itself on main (https://github.com/astral-sh/rye/pull/1125) so `powershell` is now correct (as of next release).

jimkring:
Thanks @charliermarsh.

charliermarsh:
I'll try to do a release soon so we don't have this discrepancy sitting around.

jimkring:
Yes, this impacts the getting started / first impression experience for new
users.

PR #1168, opened 2024-06-24
fix Windows Shell Completion error

OnismV:


charliermarsh:
Sorry, this was already fixed on `main` (we fixed the setting rather than the documentation). I will cut a release today to get it out.

PR #1183, opened 2024-06-30
Pass keyring-provider to uv sync and install

emarsden-iso:
When calling `uv pip {install,sync}`, the `keyring-provider` argument doesn't get passed, even though `uv` supports these arguments for both.

We already have the functionality implemented for `uv pip compile`, so just extend the support to these.

Tested locally for both commands, seems to work fine.


charliermarsh:
Thanks! Will review.

charliermarsh (approved):

PR #1192, opened 2024-07-04
Use musl build for ARM Linux

charliermarsh:
## Summary

uv's dynamically-linked ARM Linux build is only compatible with `manylinux_2_28` (we can't build Ring for earlier versions). You can see that we [dual-tag](https://pypi.org/project/uv/#files) the musl build for this reason as being `manylinux_2014`-compatible. This PR moves the ARM build to use the musl build, which should resolve the compatibility issues with `manylinux_2_17`.

Closes https://github.com/astral-sh/rye/issues/1187.


charliermarsh:
@konstin -- Can you please review?

konstin (approved):

PR #1201, opened 2024-07-08
Update Docs Description for Download Options

MichaelYochpaz:
Was just downloading Rye to try it out and got a bit annoyed by the confusing download links (like `x64 Intel Windows`  - I have an AMD CPU, and it obviously doesn't matter), so I updated them to more standard terms, although there are several options to update them to, so feel free to ask for / make changes).

I also think it might be a good idea to consider dropping the Windows x86 (32-bit) builds, as Windows 11 doesn't even support x86, and Windows 10 EOL is next year (plus, I doubt there are many Windows 10 users using the 32-bit version. Especially developers), and it just confuses newcomers with an additional option to pick from.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ The installation takes just a minute:
 
 * **Windows:**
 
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):
This looks reasonable to me, thanks.

charliermarsh:
I thought ARM Windows was pretty rare :)

PR #1203, opened 2024-07-08
Publish: Allow empty token, so twine uses keyring

emarsden-iso:
Similar in spirit to https://github.com/astral-sh/rye/pull/759, but less ambitious.

This just makes rye accept empty-string `token` values. In such a case, rye just doesn't pass a `--password` value onto `twine`. This is valid from twine's perspective; in this case twine will fall back to using [its keyring support](https://twine.readthedocs.io/en/stable/#keyring-support).

### DIFF
--- a/rye/src/cli/publish.rs
+++ b/rye/src/cli/publish.rs
@@ -252,7 +252,6 @@ fn secret_from_token(
         let maybe_encrypted = maybe_encrypt(&secret, yes)?;
         let maybe_encoded = maybe_encode(&secret, &maybe_encrypted);
         repo_credentials["token"] = Item::Value(maybe_encoded.expose_secret().into());
-
         Ok(Some(secret))
     }
 }
--- a/rye/src/cli/publish.rs
+++ b/rye/src/cli/publish.rs
@@ -77,7 +77,7 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
     // c. Otherwise prompt for token and provide encryption option, storing the result in credentials.
     let repository = &cmd.repository;
     let mut credentials = get_credentials()?;
-    credentials
+    let repo_credentials = credentials
         .entry(repository)
         .or_insert(Item::Table(Table::new()));
 
@@ -85,9 +85,8 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
         Some(url) => url,
         None => {
             let default_repository_url = Url::parse("https://upload.pypi.org/legacy/")?;
-            credentials
-                .get(repository)
-                .and_then(|table| table.get("repository-url"))
+            repo_credentials
+                .get("repository-url")
                 .map(|url| match Url::parse(&escape_string(url.to_string())) {
                     Ok(url) => url,
                     Err(_) => default_repository_url.clone(),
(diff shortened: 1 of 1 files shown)


rye/src/cli/publish.rs
```diff
@@ -77,17 +77,16 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
     // c. Otherwise prompt for token and provide encryption option, storing the result in credentials.
     let repository = &cmd.repository;
     let mut credentials = get_credentials()?;
-    credentials
+    let repo_credentials = credentials
```
emarsden-iso, line 80: This is a bit tangential, but this lets us pass less info to `secret_from_token`, just the repo credentials rather than the whole thing.

Happy to revert if unwanted, but it does remove a bunch of extra `gets` to this value, and gives us a proof that the key exists, instead of having to handle the `Option` returned by `.get(repository)`

rye/src/cli/publish.rs
```diff
@@ -103,50 +102,35 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
...
-        let maybe_encrypted = maybe_encrypt(&secret, cmd.yes)?;
-        let maybe_encoded = maybe_encode(&secret, &maybe_encrypted);
-        credentials[repository]["token"] = Item::Value(maybe_encoded.expose_secret().into());
-
-        secret
-    } else if let Some(token) = credentials
-        .get(repository)
-        .and_then(|table| table.get("token"))
+    let maybe_secret = if let Some(token) = cmd.token {
+        secret_from_token(token, cmd.yes, repo_credentials)?
```
emarsden-iso, line 113: Made this helper to deduplicate the common logic between this cmdline case and the prompt case below

rye/src/cli/publish.rs
```diff
@@ -255,6 +240,22 @@ fn maybe_decrypt(secret: &Secret<String>, yes: bool) -> Result<Secret<String>, E
     bail!("failed to decrypt")
 }
 
+fn secret_from_token(
+    token: String,
+    yes: bool,
+    repo_credentials: &mut Item,
+) -> Result<Option<Secret<String>>, Error> {
+    if token.is_empty() {
+        Ok(None)
```
emarsden-iso, line 249: this is the key bit. instead of bailing on the empty string, we return a `None` for the secret

rye/src/cli/publish.rs
```diff
@@ -157,10 +141,11 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
         .args(files)
         .arg("--username")
         .arg(username)
-        .arg("--password")
-        .arg(token.expose_secret())
         .arg("--repository-url")
         .arg(repository_url.to_string());
+    if let Some(secret) = maybe_secret {
```
emarsden-iso, line 146: the other key bit. we only add a `password` if a secret got created, ie if there was a non-empty string (or some credential looked up from file)

emarsden-iso:
@charliermarsh @mitsuhiko would you be able to take a look at this when you get a chance? happy to amend the approach as needed

PR #1227, opened 2024-07-13
Emit warning when asked to build a virtual proj

nazq:


rye/src/cli/build.rs
```diff
@@ -69,6 +69,10 @@ pub fn execute(cmd: Args) -> Result<(), Error> {
     for project in projects {
         // skip over virtual packages on build
         if project.is_virtual() {
+            warn!(
+                "skipping virtual package {}",
+                style(project.normalized_name()?).cyan()
+            );
```
charliermarsh, line 75: I wonder if we should only do this when _all_ projects are virtual?

nazq: Great suggestion

nazq: pushed an update

charliermarsh (approved):

charliermarsh:
Thanks!

PR #1250, opened 2024-07-20
Respect `.python-version` in Rye builds

charliermarsh:
## Summary

Today, `rye build` uses the bootstrapped (or "self") environment, which means built wheels reflect the version from the bootstrapped environment, rather than the user's Python preference.

This PR changes `build` to instead create an ephemeral build environment based on the requested Python version (then install and run `build` in that environment).

Creating and populating that ephemeral build environment is extremely fast with uv; there's virtually no overhead.

Closes https://github.com/astral-sh/rye/issues/1152.
Closes https://github.com/astral-sh/rye/issues/1248.

## Test Plan

- Ran `../rye/target/debug/rye build` from a directory with `.python-version` of `3.11`.
- Verified that Python 3.11 was used.

### DIFF
--- a/rye/src/bootstrap.rs
+++ b/rye/src/bootstrap.rs
@@ -35,7 +35,7 @@ pub const SELF_PYTHON_TARGET_VERSION: PythonVersionRequest = PythonVersionReques
 
 const SELF_VERSION: u64 = 20;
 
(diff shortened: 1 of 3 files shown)


b0o:
Awesome! Seems to be working as expected.

PR #1326, opened 2024-08-11
Fix a few regex inefficiencies

InSyncWithFoo:
Explanations:

* `.*?$` is the same as `.*`: Being lazy at the end is the same as being greedy.
* `[^\-]` is the same as `[^-]`: The `-` here cannot be considered a part of a range.
* The dot in `user.(name|email)` acts as a metacharacter.
* `(?:(via)|(?:via (.*?))|(?:  (.*?)))$` is equivalent to `(?:via$|via (.*)|  (.*))`.
    Consequently, a line below was changed from `.get(2).or_else(|| m.get(3))` to `.get(1).or_else(|| m.get(2))`.


rye/tests/test_self.rs
```diff
@@ -52,8 +52,8 @@ fn test_version() {
 
     let mut settings = Settings::clone_current();
     settings.add_filter(r"(?m)^(rye )\d+\.\d+\.\d+?$", "$1[VERSION]");
```
charliermarsh, line 54: Should the `?` here to be removed too?

PR #1328, opened 2024-08-12
Allow specifying script commands relative to the project root

brandon-avantus:
Command paths with a leading `./` have the project root prepended, which allows scripts in the project tree to be used with `rye run` even when the current directory is not the project root.


rye/src/cli/run.rs
```diff
@@ -106,6 +106,11 @@ fn invoke_script(
                     .map(OsString::from)
                     .chain(args.into_iter().skip(1))
                     .collect();
+
+                let path = std::path::Path::new(&args[0]);
+                if path.starts_with("./") {
+                    args[0] = pyproject.root_path().join(path).into();
```
charliermarsh, line 112: I think we should just _always_ join this, since if `path` is absolute, joining just resolves to `path`.

PR #1351, opened 2024-08-25
rye list: Never modify or delete the virtualenv

### LINKED ISSUE #940: rye list overwrites .venv
### Steps to Reproduce

1. Enter a non-rye managed project that has pyproject.toml and .venv
2. rye list

(for example with PDM: `mkdir project; cd project; pdm init -n; pdm add jinja2; pdm list; rye list; pdm list;` )

### Expected Result

rye list should error or make a best effort to list the packages in the project.

rye list should not wipe out .venv directories that rye is not managing. rye list should preferably never do any modifications at all, it's a query command.

### Actual Result

It overwrites .venv with a fresh venv (created with uv venv), wiping anything that was there before.

```
Using Python 3.10.12 interpreter at: .venv/bin/python
Creating virtualenv at: .venv
Activate with: source .venv/bin/activate
```

### Version Info

rye 0.31.0
commit: 0.31.0 (72e239b8f 2024-03-22)
platform: linux (x86_64)
self-python: cpython@3.12.2
symlink support: true
uv enabled: true

### Stacktrace

_No response_

### LINKED ISSUE #1352: rye list overwrites (managed) venv if python version does not match
### Steps to Reproduce

In this case, rye list overwrites the venv. In this test case, the project is rye managed, but it's still surprising that rye list modifies the project.

1. `mkdir project; cd project; rye init -p 3.12; rye sync; rye pin 3.11.7; rye list; cd -`

Using uv 0.3.3

### Expected Result

`rye list` keeps the existing virtualenv

### Actual Result

`rye list` overwrites `.venv` with a new empty venv.

We have tool.rye.managed=true but are still surprised that rye list was destructive.

### Version Info

rye 0.39.0
commit: 0.39.0 (bf3ccf818 2024-08-21)
platform: linux (x86_64)
self-python: cpython@3.12.4
symlink support: true
uv enabled: true

### Stacktrace

_No response_

bluss:
Previously, in *some* situations, `rye pin <pythonversion>; rye list` could lead to rye deleting the existing virtualenv.

I suggest that `rye list` should be an interrogator and never modify any project. With this change, it lists its own and non-rye virtualenvs (same behaviour as uv pip freeze) but should never create, recreate or overwrite environments (own or non-rye).

Fixes #1352


bluss:
Making draft while figuring out some questions in #940

charliermarsh (approved):
This makes sense to me, thanks. I just tweaked some of the names.

PR #1383, opened 2024-09-14
fix: unexpected abort when symlink to non-regular file

fourdim:
`std::fs::copy` will return an error when
- from is neither a regular file nor a symlink to a regular file (folder).
https://doc.rust-lang.org/std/fs/fn.copy.html

Error screenshot:
image

Windows part is not tested.

### DIFF
--- a/rye/src/utils/mod.rs
+++ b/rye/src/utils/mod.rs
@@ -453,6 +453,39 @@ pub fn copy_dir<T: AsRef<Path>>(from: T, to: T, options: &CopyDirOptions) -> Res
                 fs::create_dir_all(&destination)
                     .path_context(&destination, "failed to create directory")?;
                 copy_dir(entry.path(), destination, options)?;
+            } else if entry.file_type()?.is_symlink() {
+                let target = fs::read_link(&entry_path)
+                    .path_context(&entry_path, "failed to read symlink target")?;
+                #[cfg(unix)]
+                {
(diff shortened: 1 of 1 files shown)


rye/src/utils/mod.rs
```diff
@@ -453,6 +453,39 @@ pub fn copy_dir<T: AsRef<Path>>(from: T, to: T, options: &CopyDirOptions) -> Res
                 fs::create_dir_all(&destination)
                     .path_context(&destination, "failed to create directory")?;
                 copy_dir(entry.path(), destination, options)?;
+            } else if entry.file_type()?.is_symlink() {
+                let target = fs::read_link(&entry_path)
+                    .path_context(&entry_path, "failed to read symlink target")?;
+                #[cfg(unix)]
+                {
+                    if target.is_absolute() && target.starts_with(from) {
+                        symlink_file(target.strip_prefix(from).unwrap(), &destination)
```
charliermarsh, line 462: Why is the `strip_prefix` here necessary?

fourdim: For example:
If we have a folder symlink:
```
/project/folder -> /project/folder2
```
And we want to copy it to `/tmp`,
It is apparently we don't want this:
```
/tmp/folder -> /project/folder2
``` 
This one is more reasonable, isn't it?
```
/tmp/folder -> ./folder2
```

PR #1433, opened 2024-11-18
Update Dockerfile example, need to COPY README.md and pyproject.toml

Athroniaeth:
Going back to the Dockerfile example, I saw that `uv pip install ...` required a pyproject.toml and README.md file (generated by `rye init`). You therefore need to copy them into the Dockerfile (which is not specified).

When i don't copy pyproject.toml
pyproject

When i don't copy README.md
image

### DIFF
--- a/docs/guide/docker.md
+++ b/docs/guide/docker.md
@@ -22,7 +22,7 @@ This guide requires some familiarity with Docker and Dockerfiles.
     RUN pip install uv
 
     WORKDIR /app
-    COPY requirements.lock ./
+    COPY README.md pyproject.toml requirements.lock ./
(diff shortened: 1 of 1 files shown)


charliermarsh:
Hmm. If your project is virtual, I don't think this should be required, since we shouldn't be building it. Are you using a non-virtual project?

Athroniaeth:
> Hmm. If your project is virtual, I don't think this should be required, since we shouldn't be building it. Are you using a non-virtual project?

Hello, I reviewed my configuration files and just realized that I misconfigured them. Apologies for the inconvenience; I'm closing the PR.

PR #1437, opened 2024-11-22
Ignore freethreaded CPython builds

### LINKED ISSUE #1436: Rye uses indygreg's freethreaded builds
e.g. https://github.com/astral-sh/rye/blob/7b8b1016ad6b0708639bfeb43c65911d40da9e0d/rye/src/sources/generated/python_downloads.inc#L104-L108

It's unclear whether this was intended as this was introduced automatically by https://github.com/astral-sh/rye/pull/1406. Per <https://docs.python.org/3/whatsnew/3.13.html> "support for running in a [free-threaded mode](https://docs.python.org/3/whatsnew/3.13.html#whatsnew313-free-threaded-cpython)" is "experimental".

Freethreaded CPython builds can cause issues e.g. https://github.com/pydantic/pydantic-core/issues/1548 that don't happen with the regular builds.

Would it make sense to download non-freethreaded builds by _default_ (and maybe let folks opt into freethreaded builds, e.g. by requesting a `3.13.0t` build rather than a `3.13.0`)?

Alexis-D:
Fixes #1436

This seemed like the simplest fix to ignore freethreaded builds.

Couple other thoughts/things that could happen later:

* this won't help with folks who already downloaded a 3.13.0 CPython (they'll need to `rye toolchain remove cpython@3.13.0` first, and then `fetch` it again)
* `PlatformTriple#flavor` should probably be a `set[str]`; so `freethreaded+pgo+lto-full` would become `{'freethreaded', 'pgo', 'lto'}` (tbd what to do with `full`)


zanieb (approved):
Awesome thank you!

PR #1455, opened 2025-01-16
Add riscv64-linux for find_downloads.py

Xeonacid:
https://github.com/astral-sh/python-build-standalone/releases/tag/20250115 added riscv64-linux.


Xeonacid:
The clippy failure should be unrelated to the PR.

PR #1476, opened 2025-07-21
Retire Rye and add a uv migration guide

geofft:
https://lucumr.pocoo.org/2024/8/21/harvest-season/

> Domination is a goal because it means that most investment will go
> into one stack. I can only re-iterate my wish and desire that Rye (and
> with it a lot of other tools in the space) should cease to exist once
> the dominating tool has been established. For me uv is poised to be
> that tool. It's not quite there today yet for all cases, but it will
> be in no time, and now is the moment to step up as a community and
> start to start to rally around it. That doesn't mean that this tool
> will be the tool forever. Things come and go and maybe there is a
> future for some other tool.
>
> But today I'm looking forward to the moment when there will be a final
> release of Rye that is no remaining functionality other than to just
> largely alias to uv, that retires Rye specific functionality and
> migrates you over to uv.

This does not do a final release - adding a compatibility layer for uv and an automatic migration is likely to be buggy and incomplete, and users are better served by changing the command they type to `uv`. Instead, it adds a migration guide based on the discussion thread (https://github.com/astral-sh/rye/discussions/1342), documenting things that are not immediately obvious to a user swapping out `rye` for `uv`.

### DIFF
--- a/docs/guide/uv.md
+++ b/docs/guide/uv.md
@@ -4,7 +4,7 @@ Rye is no longer developed. We recommend that Rye users migrate to [uv](https://
 
 Rye and uv have similar philosophies, CLIs, and user experience. Both are project management tools that follow Python standards wherever possible, so migrating your workflows should be easy. Start with [uv's getting started guide](https://docs.astral.sh/uv/getting-started/).
 
-This document has guidance on how to migrate specific Rye features to uv, targeted at uv 0.8.0 / July 2025. If you're on an older version of uv, please upgrade. (If you are reading this page in the future, check the linked issues below to see if even more features have been implemented in uv!)
+This document has guidance on how to migrate specific Rye features to uv, targeted at uv 0.8 / July 2025. If you're on an older version of uv, please upgrade. (If you are reading this page in the future, check the linked issues below to see if even more features have been implemented in uv!)
 
 ## Migrating existing projects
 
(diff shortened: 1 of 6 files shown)


geofft:
[Rendered website](https://kerberos.club/tmp/rye/)

hugovk:
Thanks for this migration guide, the "`rye fmt` and `rye test`" section was especially useful.

One thing that was missing, replace `rye lint` with something like `uv run ruff check`.

geofft:
Great, glad it was useful! Added a section on that (and updated the preview rendering above so the links work right).

docs/guide/uv.md
```diff
@@ -0,0 +1,83 @@
...
+## Migrating existing projects
+
+If you don't have `[tool.rye]` sections in your `pyproject.toml` file, your project should work out of the box in uv with no further changes.
+
+If you have `[tool.rye]` sections in your `pyproject.toml` file, you will need to convert them to `[tool.uv]` sections. Most of the time, all you need to do is rename the sections.
+
+A few configuration keys are slightly different but easy to migrate:
+
+* `tool.rye.universal` and `tool.rye.generate-hashes` are under the `tool.uv.pip` namespace, i.e., `tool.uv.pip.universal` and `tool.uv.pip.generate-hashes`.
+* `tool.rye.lock-with-sources` corresponds to `tool.uv.no-sources` but with the sense inverted, i.e., convert `lock-with-sources = false` to `no-sources = true`.
```
zanieb: "sense inverted" seems confusing

docs/guide/uv.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Python installations and shims
+
+Both uv and Rye support installing the same managed Python releases from [python-build-standalone](https://github.com/astral-sh/python-build-standalone) (now under Astral stewardship), so you should find Python behavior to be highly compatible.
+
+Rye supports a [global Python shim](https://rye.astral.sh/guide/shims/) that sets up the `python` and `python3` commands to run a managed version of Python. While this is highly convenient, it does risk breaking existing software on your computer that expects the OS version of Python.
+
+As of uv 0.8.0, `uv python install` will install a `python3.x` command for the latest stable managed Python. If you would like to install `python` and `python3` commands too. you can use `uv python install --preview`. Note that this is _not_ a shim, just a direct link to the Python installation; it does not pick up dependencies from the current project. This behavior may change in the future; [astral-sh/uv#6265](https://github.com/astral-sh/uv/issues/6265) tracks uv adding a Python shim.
```
zanieb: It's `--default` not `--preview` here

zanieb (approved):

docs/guide/uv.md
geofft: How about
```md
* `tool.rye.lock-with-sources` corresponds to `tool.uv.no-sources` but the Boolean value of the setting is inverted, i.e., convert `lock-with-sources = false` to `no-sources = true`.
```
?

zanieb: That works for me, though I'd nit and not capitalize "Boolean".

geofft: I've always thought of it as a proper noun (after George Boole, similar to "Euclidean distance", "Hamiltonian", etc.), but huh, the Rust docs seem to use lowercase and the CPython docs use both styles. Wikipedia uses uppercase.

hugovk: Added to the CPython style guide just yesterday:
https://devguide.python.org/documentation/style-guide/#specific-words

It definitely varies. 

See also https://english.stackexchange.com/questions/13762/are-there-examples-of-terms-named-after-a-person-that-are-no-longer-capitalized for a whole list of eponyms that are no longer capitalised.

